PR #62, opened 2013-12-18
update to gnu3

ghost:


StackNeverFlow:
Updating to GNU3 is not makeable, you need the agreement of ever code editor.

istiak101:
Ban Israel from Linux kernel

PR #74, opened 2014-02-25
Update resize.c

thedj21:
Update resize.c


guglovich:
Please tell me full example for f2fs resize. The Internet is not found, and the standard instruction I did not understand.

PR #76, opened 2014-03-05
Fix section 3.8 number

ruckuus:


Elizafox:
Useless. And Linus doesn't accept PR's, to add insult to injury.

swarren:
Eliza, I would not say that fixing documentation is useless. That's a little demeaning.

FWIW, this bug is fixed upstream already by 49d063cb3532 "proc: show mnt_id in /proc/pid/fdinfo".

PR #81, opened 2014-03-18
facebook functionality

raincoats:
allows users to login to linux with facebook


istiak101:
Ban Israel from Linux kernel

PR #17, opened 2012-05-11
Add support for AR5BBU22 [0489:e03c]

reejk:

PR #90, opened 2014-04-08
Update ratelimit.c

renweihub:
fix a bug in ratelimit, if the "begin" doesn't update at the same time with printed, the print will start from 1 in loops except for the first time, lets to only 9 logs suppressed, but not 10 as expected.


Elizafox:
@renweihub [Linus doesn't accept PR's on github](https://github.com/torvalds/linux/pull/17#issuecomment-5654674)

samcn:
Omg fuck off Elizabeth Myers

swarren:
samcn, that kind of language really isn't appropriate.

ariccio:
> samcn, that kind of language really isn't appropriate.

(That's a right reserved for Linus, and Linux only)

PR #91, opened 2014-04-23
Fix the double-barreled homophone attack

Ulexus:
This change is [Reviewable](https://reviewable.io/reviews/torvalds/linux/91)


Elizafox:
A really minor docs change that's slightly irritating, and should be upstreamed, but there's just one problem. [Linus doesn't accept pull requests on github](https://github.com/torvalds/linux/pull/17#issuecomment-5654674), so send them to the appropriate maintainer.

PR #92, opened 2014-04-23
Add auto detection of the ds1307 device on the I2C

maestrx:
Code should add instructions to kernel to auto-detect the presence of ds1307 device on I2C bus on address 0x68 and in case it is found auto assign it. The main drive is to make sure that devices (like Pi) do have correct time before init process is started thus make sure that all daemons have correct time during startup phase. I'm pretty sure that the code added is wrong, jsut don t knwo whom to contact to get help to add this functionality :) I've reused parts from the DOC: https://www.kernel.org/doc/Documentation/i2c/instantiating-devices


Elizafox:
[Linus doesn't accept pull reuqests on GitHub](https://github.com/torvalds/linux/pull/17#issuecomment-5654674).

PR #94, opened 2014-05-23
Update saa7134-cards.c

AnTAVR:
fix for AVERMEDIA STUDIO 507UA


Elizafox:
@AnTAVR Linus doesn't accept pull requests from GitHub. Consult https://github.com/torvalds/linux/blob/master/Documentation/HOWTO instead.

PR #97, opened 2014-05-30
USB: cp210x: Add 8998 (KCF PRN Device)

sonofaforester:
Adds support for the KCF Technologies PRN Device in the cp210x usb serial driver


Elizafox:
Linus doesn't accept PR's on github.

PR #116, opened 2014-08-23
Update names.c

ghost:
Hello.
While making a program using usbip, I found a bug.
When the variable pool_head was free()ed, it has dangling pointer. So, after names_free() called, we can't call names_free() again.
Thanks for reading. 
(Sorry for not fluent English)


Elizafox:
Fantastic, but Linus doesn't accept pull requests.

PR #119, opened 2014-09-14
Create voh98

ghost:


Elizafox:
@Johvoo Not only is this totally worthless, Linus doesn't accept pull requests from GitHub. Sorry!

PR #128, opened 2014-10-24
Eliminate two duplicated lines codes to enhance the xt_hashlimit.c of ne...

gfreewind:
There are two duplicated lines of hashlimit_mt. Now I enhance it to eliminate the duplicated codes.


Elizafox:

istiak101:
Ban Israel from Linux kernel

PR #135, opened 2014-11-14
Small comment correction

foodprocessor:
The time field in blk_io_trace is in nanoseconds. Correct the comment from microseconds to nanoseconds.
The field was once in microseconds, but was changed to nanoseconds in commit 6c051ce0307526adec32a847f0daa1af2124f0a9 on March 25, 2009. Update the comment to match a 5-year-old reality.


septyaman:
:(

PR #136, opened 2014-11-15
Update xpad.c

linbingchen007:
the abs_y and the abs_ry should negate the val and add 1 because it's the complement of the origin variable


Elizafox:

PR #139, opened 2014-11-26
Minor improvement for smsc95xx netusb driver performance.

SiarheiVolkau:
Reduce number of memcpy's by 1-2 improve transmit performance by 2-4%, or reduce cpu usage on a comparable value.


Elizafox:

PR #143, opened 2014-11-30
Update abdac.c

Amitabha2001:
Some mini changes of int to void.


Elizafox:

PR #150, opened 2015-01-06
Replace gzip/bzip2 with xz in README

felixphew:
Kernel tarballs are only compressed with xz now, gzip and bzip2 are redundant.


joshumax:
Linus doesn't accept PRs on GitHub.

felixphew:
Where should I file it? I didn't see any pull request mechanism on kernel.org, and it was only a minor change (doesn't even change code)

PR #155, opened 2015-01-17
Moving index in a more precise scope.

Triton64:


logangorence:
Pull requests are not accepted on GitHub for Linux.

joshumax:
You do realize that C doesn't allow type declarations inside a for loop except in C99...right? If you tried to compile the kernel with these changes you should have noticed the big, huge, unavoidable compilation error that would appear.

Plus Linux doesn't accept PRs on GitHub...

ariccio:
Where's the type declaration inside a loop?? Not seeing it. 

[and, yes, Linus __really__, _really_, doesn't accept pull requests on GitHub](https://github.com/torvalds/linux/pull/17#issuecomment-5654674)

PR #160, opened 2015-02-11
mali_drv.h includes drm_sman.h 

amegz:
There no drm_sman.h in linux kernel anymore


immibis:

PR #181, opened 2015-04-26
Genius MousePen i608x patch fix

Erendis:
okay, now i just sucked myself in front of everyone.
then i'd like to add a few lines required for the device mentioned to work properly. thanks in advance.

### DIFF
--- a/drivers/hid/hid-kye.c
+++ b/drivers/hid/hid-kye.c
@@ -186,6 +186,108 @@ static __u8 mousepen_i608x_rdesc_fixed[] = {
 	0xC0              /*  End Collection                  */
 };
 
+/*-* unfortunately the patch does not work properly without the following lines.
+     (see http://askubuntu.com/questions/547041/pen-tablet-mousepen-doesnt-work-in-ubuntu for more information)  *-*/
+
+    /* Original MousePen i608X report descriptor size */
+    #define MOUSEPEN_I608X_2_RDESC_ORIG_SIZE  482
+
+    /* Fixed MousePen i608X report descriptor */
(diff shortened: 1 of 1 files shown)


bombless:
I like Github PR but clearly not everybody likes it :disappointed:

joshumax:
This issue has already been fixed a while ago, your HID mapping is exactly the same as the one currently in the code-base (See https://www.diffchecker.com/bkdefals), and since there's no additional commits tied to your PR it seems that your fix is simply an unused array (See https://github.com/Erendis/linux/blob/master/drivers/hid/hid-kye.c#L558).

If you are in a bug-fixing mood though, the LKML receives several bug reports each day and I'm sure any patches that correctly fix them would be appreciated.

Erendis:
it has 4 differences, although the second one is only a _2 suffix:

https://github.com/Erendis/linux/blob/master/drivers/hid/hid-kye.c#L193
https://github.com/Erendis/linux/blob/master/drivers/hid/hid-kye.c#L196
https://github.com/Erendis/linux/blob/master/drivers/hid/hid-kye.c#L236
https://github.com/Erendis/linux/blob/master/drivers/hid/hid-kye.c#L240

Erendis:
the line you mentioned is already there indeed, but this block is also needed for the driver to work properly with full drawing area (ymmv)

joshumax:
@Erendis It seems the patch you're trying to send as a PR has already been sent to the LKML (http://www.spinics.net/lists/linux-input/msg31764.html). But I'm still confused, exactly where in your Pull Request is `mousepen_i608x_2_rdesc_fixed` referenced?

Erendis:
yeah, right. cool. thanks!

PR #182, opened 2015-05-06
Another Hardware ID That i want to "just work".

deddokatana:
Im Just a Python Programmer :+1:


ecnepsnai:
Well if you want things to "Just work" you should follow the proper procedure for submitting patches: https://github.com/torvalds/linux/blob/master/Documentation/SubmittingPatches

PR #204, opened 2015-09-08
Fixes issue preventing DMA being used

rlbartle:
Fixes issue preventing DMA being used whenever the rx/tx_buf is not word aligned.


rumpelsepp:
https://github.com/torvalds/linux/blob/master/Documentation/SubmittingPatches

zoutianya521314:
Excuese me ,guys, I have a problem:
    now, I use xilinx zynqmp chip, which is made up of PS and PL, that is to say ARM and FPGA. 
and the linux sys is running on ARM(system transplant is done), the data stored in the ddr4 of PL should be transfer to PS by using the DMA(I understand truly), there are two dma driver files named zynqmp_dma.c and xilinx_dma.c ,can I use them? if so, how to call the driver? just modify the device-tree and write app program,seems a little diffcult.
   please give me some suggestions!

PR #207, opened 2015-09-17
Add the missing PCIID for MSI (maybe not only) R7 370

mradermaxlol:
There's a missing PCIID for MSI R7 370 Armor 2X (probably some other cards). Without it the screen will turn black (there would be signal, but system hangs if it happens) or just become filled with vertical white stripes - note that it happens at the modesetting stage. There's a way to make it work without this patch by disabling dpm (dpm=0) and/or disabling modesetting (nomodeset) (note that disabling modesetting will make your system work for a while, but it will hang after a few minutes). Adding this line/ID will make the GPU function as expected, without functionaluty loss/instability.


Airblader:
The kernel doesn't do PRs.

PR #209, opened 2015-09-29
Added bc in the list of dependencies to have an exhaustive list.

blem3:
I created a docker container based on the list of dependencies and it was missing bc.
I spotted out it is later in the documentation but it would be more convenient to have it in the list as well as it seems to be the only one missing when using the default configuration on x86_64.


blem3:
Will sublit through the official process.

PR #242, opened 2016-01-09
Decreasing the size of radix_tree_node

mzer0-yu:
The count in radix_tree_node does NOT need to be unsigned int but unsigned short is enough. For the disability of calculating logarithm in compile time, original authors defined RADIX_TREE_HEIGHT_SHIFT as RADIX_TREE_MAX_PATH + 1, but it should be log2(RADIX_TREE_MAX_PATH) + 1 indeed. I noticed that and, after calculating I am aware of unsigned long is not suitable for count but unsigned short. Between, I knew there is an alignment of radix_tree_node, so I cannot tell if the size of radix_tree_node will actually decrease.

### DIFF
--- a/include/linux/radix-tree.h
+++ b/include/linux/radix-tree.h
@@ -76,17 +76,39 @@ static inline int radix_tree_is_indirect_ptr(void *ptr)
 #define RADIX_TREE_MAX_PATH (DIV_ROUND_UP(RADIX_TREE_INDEX_BITS, \
 					  RADIX_TREE_MAP_SHIFT))
 
+/* log2(RADIX_TREE_INDEX_BITS) will not above 10. */
+#if sizeof(unsigned long) == 16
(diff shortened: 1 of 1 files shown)


quant67:
Have you compare the values of sizeof(struct radix_tree_node) between old and new version? I guess it change nothing unless a packed attribute is there. 

I suggest you that submit path on a right way: https://github.com/torvalds/linux/blob/master/Documentation/SubmittingPatches

> Linus, in particular, will not pull from public hosting sites like GitHub in the absence of a signed tag.

ancaemanuel:
Tutorial: https://www.youtube.com/watch?v=LLBrBBImJt4
Setting git send mail: https://coderwall.com/p/dp-gka/setting-up-git-send-email-with-gmail
Use get_maintainer.pl
Use the mailing lists.
Close this please, and tell to others.

PR #258, opened 2016-02-21
added support for goobay Active USB 3.0 Data Link Cable 

danielkucera:
I added support for PL-27A1 just by defining new vendor/device id to current plusb driver. The chipset is identical to PL-25A1 except it support USB3.0 interface.


parrotgeek1:
https://shkspr.mobi/blog/2014/04/submitting-trivial-linux-kernel-patches/

avertis:
Please add te path to the mainline.

PR #261, opened 2016-03-01
Added busybox as extra failsafe init

jcjordyn130:
Most If not all distributions include a static version of busybox in there initramfs. So I added busybox ash as a failsafe init.

### DIFF
--- a/init/main.c
+++ b/init/main.c
@@ -969,6 +969,8 @@ static int __ref kernel_init(void *unused)
 	if (!try_to_run_init_process("/sbin/init") ||
 	    !try_to_run_init_process("/etc/init") ||
 	    !try_to_run_init_process("/bin/init") ||
+	    !try_to_run_init_process("/bin/busybox ash") ||
+	    !try_to_run_init_process("/bin/bash") ||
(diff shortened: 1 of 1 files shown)


philipperemy:
Tutorial: https://www.youtube.com/watch?v=LLBrBBImJt4
Setting git send mail: https://coderwall.com/p/dp-gka/setting-up-git-send-email-with-gmail
Use get_maintainer.pl
Use the mailing lists.
Close this please, and tell to others.

immibis:
Apart from that, if I were you I'd make sure `try_to_run_init_process` can actually take arguments like that.

jcjordyn130:
You are right, I just haven't had enough time to compile it.

parrotgeek1:
@immibis it can't, it only takes the path to program and runs it with empty argv, so this is wrong. you would have to change it to set argv[0] to "ash" in that special case, which becomes too busybox-specific to be accepted to mainline

nishankbhati:
Okay. I apologize.

philipperemy:
@nishankbhati can you close it?

PR #290, opened 2016-05-19
text  formatting

renancunha33:
Just a simple format correction.


diegoviola:
@renancunha33 Linus doesn't accept GitHub PR. You need to send a patch to corbet@lwn.net (documentation maintainer) and CC linux-kernel@vger.kernel.org and trivial@kernel.org.

PR #293, opened 2016-05-27
Adjust the speed on the TUN/TAP interface

aetaric:
SPEED_10 is almost unheard of in modern Networking. Since the TUN/TAP interface can exceed this value it breaks tools like sar that use the port's speed to calculate interface utilization. SPEED_100 is a more realistic number that you might see on a modern network, though SPEED_1000 is becoming more common now.


aetaric:
Submitted via signed-off patch over email... Closing this.

PR #296, opened 2016-06-11
The description shouldn't contain a example, when the example is directly below.

tigoesnumb3rs:
Nothing big, but seems more consistent to me this way.


smclt30p:
Why do people keep doing this? Linus does NOT pull from GitHub!

FadeMind:
@tigoesnumb3rs 

and:
https://www.reddit.com/r/linux/comments/4ltsxe/dont_submit_changes_for_linux_kernel_via_github/

tigoesnumb3rs:
> Why do people keep doing this? Linus does NOT pull from GitHub!

Probably because they don't know.

Thanks for the information @FadeMind! Your comment was very helpful!

PR #308, opened 2016-07-27
Grammar mistake

Jonatino:


vikaskyadav:
Where is the mistake?

guan2317:
I am confused

mueddib:
The update is must, not use "if name doesn't have". Actually, "if the name doesn't".

https://github.com/torvalds/linux/pull/308/files#diff-6e489a0861c45e4484b7ea5b409e223e

NogradThGin:
Is this really useful to create a pull request only for that such thing ?

PR #315, opened 2016-08-18
Added usb-device desriptor for Sagem USB-Serial Controller

krajcovic:
Changes to be committed:
	modified:   drivers/usb/class/cdc-acm.c


krajcovic:
dmesg
[ 1380.309578] usb 1-1.3: new full-speed USB device number 5 using dwc_otg
[ 1380.412944] usb 1-1.3: New USB device found, idVendor=079b, idProduct=0028

PR #318, opened 2016-09-03
Change "M" to "N"

AppIns:
Someone antecedently hit M I think

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -9,7 +9,7 @@
 			Linus
 ----------
 
-M: Matt Mackal
+N: Matt Mackal
 E: mpm@selenic.com
 D: SLOB slab allocator
 


samangh:
This repo is mirror-only, nothing will be accepted here for legitimate reasons. See:

[#17 (comment)](https://github.com/torvalds/linux/pull/17#issuecomment-5654674)

    I don't do github pull requests.

    github throws away all the relevant information, like having even a
    valid email address for the person asking me to pull. The diffstat is
    also deficient and useless.

    Git comes with a nice pull-request generation module, but github
    instead decided to replace it with their own totally inferior version.
    As a result, I consider github useless for these kinds of things. It's
    fine for hosting, but the pull requests and the online commit
    editing, are just pure garbage.

    I've told github people about my concerns, they didn't think they
    mattered, so I gave up. Feel free to make a bugreport to github.

    Linus

AraHaan:
Well how to make a bug report when the issues Tab is not set on the Repo. consider Settign ut up with 1 check in the Repo's settings.

DmitryHetman:
Install Gentoo, here you can have all python versions at the same time.

battlesnake:
Python 3.5 as default would break tonnes of install scripts, there's stuff like virtualenv for dealing with multiple versions

battlesnake:
Also, I'm not sure what Python or UI for connecting to routers has to do with the kernel....

xuhdev:
@AraHaan Are you really talking about the kernel? Check out [here](http://www.tldp.org/LDP/sag/html/gnu-or-not.html).

PR #319, opened 2016-09-06
Update bcm4709-netgear-r8000.dts

adityaxavier:
Including missing WAN GPIO and LED Switch.


adityaxavier:
Incorporate missing WAN LED and LED Switch GPIO pins.

awaismasood:
Please see:

> I don't do github pull requests.

>github throws away all the relevant information, like having even a
>valid email address for the person asking me to pull. The diffstat is
>also deficient and useless.

>Git comes with a nice pull-request generation module, but github
>instead decided to replace it with their own totally inferior version.
>As a result, I consider github useless for these kinds of things. It's
>fine for *hosting*, but the pull requests and the online commit
>editing, are just pure garbage.

>I've told github people about my concerns, they didn't think they
>mattered, so I gave up. Feel free to make a bugreport to github.

 >                   Linus

PR #323, opened 2016-09-23
Add rtc-pcf2127 as a module in the kernel config

paulcrawford:
I, and I assume many others, are using the pcf2127 chip as an rtc on the odroid c2. To use this hardware on the i2c network requires that rtc-pcf2127 be available as a module, which it is not presently. I have tested its inclusion as a module in a custom build of the kernel and it works fine. Adding it as a module in the regularly updated kernel will make upgrades much easier for all those using this chip (shield).


fuzun:
As far as I know Github pull requests don't get accepted. You may need to follow authors procedures for apply.

PR #338, opened 2016-10-18
Disable hw TSO for atl1e atheros card

anthonytex:
So maybe i'm the only one that is still using this Ethernet card in the world but after opening a bug a lot time ago my patch was not merged: this card has a broken hardware as described [here](https://bugzilla.kernel.org/show_bug.cgi?id=113501) it causes a lot of speed problem.


AntonBoch1244:
@anthonytex, Send this PR to torvalds/linux @ kernel.org repo instead of this repo, because it's mirror of torvalds/linux @ kernel.org repo.

anthonytex:
Done, thanks @AntonBoch1244

PR #350, opened 2016-11-01
Sun8i H3: i2c2 uses regs for i2c1

WereCatf:
At line 539
reg = <0x01c2b000 0x400>;
should be
reg = <0x01c2b400 0x400>;


arch/arm/boot/dts/sun8i-h3.dtsi
```diff
@@ -519,7 +519,7 @@
 
 		i2c2: i2c@01c2b400 {
 			compatible = "allwinner,sun6i-a31-i2c";
-			reg = <0x01c2b000 0x400>;
+			reg = <0x01c2b400 0x400>;
```
odykyi: pls add tests

PR #356, opened 2016-11-06
Add Microsoft Surface Pro 4 Type Cover support

jfarthing84:


jfarthing84:
Looks like Github pull requests aren't accepted.

PR #364, opened 2016-12-07
Corrected a Comment

WesCourtney44:
Changed the misspelled word "behaviour" to "behavior" on line 339 of the core.c file.

### DIFF
--- a/kernel/livepatch/core.c
+++ b/kernel/livepatch/core.c
@@ -336,7 +336,7 @@ unlock:
  * Convert a function address into the appropriate ftrace location.
  *
  * Usually this is just the address of the function, but on some architectures
- * it's more complicated so allow them to provide a custom behaviour.
+ * it's more complicated so allow them to provide a custom behavior.
  */
 #ifndef klp_get_ftrace_location
 static unsigned long klp_get_ftrace_location(unsigned long faddr)


codernator:
Seriously?
[](http://grammarist.com/spelling/behavior-behaviour/)
What release should your super-critical PR go into?

plusreed:
Even if this pull request **was** merged (which it won't, because Linus doesn't take pull requests from GitHub), it wouldn't even change the meaning of the comment because both of the words mean the exact same thing. It's only the spelling that's different from British English (behaviour) and American English (behavior).

About PRs not being accepted here, [this is how you submit patches to the kernel.](https://github.com/torvalds/linux/blob/master/Documentation/SubmittingPatches) No pull requests will be accepted here, [this is the reason why.](https://github.com/torvalds/linux/pull/17#issuecomment-5654674)

WesCourtney44:
Sorry guys, I wasn't trying to be sarcastic or anything. I'm new to github and was trying to make a very minute first contribution that wouldn't be obstructive to anything. It was more of a test, to see how I could push real content in the future.

Plusreed, thank you for your feedback I was unaware that pull requests weren't to be accepted, thanks to you I now know why. Perhaps I should have checked the British spelling spelling...oops.  :)

PR #370, opened 2016-12-30
net/bridge: fix dropping packets that moving through bridge interface

Hexta:
Fix regression introduced by commit c5136b1.
br_nf_pre_routing_finish() calls itself instead of
br_nf_pre_routing_finish_bridge().

Due to this bug reverse path filter drops packets that moving through bridge interface.


Hexta:
Close since it was fixed by https://github.com/torvalds/linux/commit/14221cc45caad2fcab3a8543234bb7eda9b540d5

PR #376, opened 2017-01-11
4.8/wb buf throttle

bm16ton:
testing


ItzLevvie:
PRs don't get accepted here.

PR #385, opened 2017-01-25
mips checksum error -- csum_tcpudp_nofold

bombhero:
If the input parameters as saddr = 0xc0a8fd60,daddr = 0xc0a8fda1,len = 80, proto = 17, sum =0x7eae049d.
The correct result should be 1, but original function return 0.


benlavender:
Linus does not take PRs on here, see either the rest of the PRs here and this:
https://github.com/torvalds/linux/pull/17#issuecomment-5663780

PR #388, opened 2017-02-02
Fix tmp reusage in net2280 driver

RazManor:
There was a discussion in the mailing list about the reuse of tmp in scan_dma_completions().
I made the changes as linus suggested, according to my original patch.
The changes works for me, hope that others will be able to test them, and merge them.

This was my logic for the patch:
1. The req->td->dmadesc equals to 0 iff:
     -- There was a transaction ending with a short packet, and
     -- The read() to read it was shorter than the transaction length, and
     -- The read() to complete it is longer than the residue.
     I believe this is true from the printouts of various cases, but I can't be positive it is correct.

2. Entering this if, there should be no more data in the endpoint (a short packet terminated the transaction). If there is, the transaction wasn't really done and we should exit and wait for it to finish entirely. That is the inner if.
    That inner if should never happen, but it is there to be on the safe side. That is why it is marked with the comment /* paranoia */. 
    The size of the data available in the endpoint is ep->dma->dmacount and it is read to tmp.
     This entire clause is based on my own educated guesses.

3. If we passed that inner if without breaking in the original code, than tmp & DMA_BYTE_MASK_COUNT== 0.
    That means we will always pass dma bytes count of 0 to dma_done(), meaning all the requested bytes were read.

4. dma_done() reports back to the upper layer that the request (read()) was done and how many bytes were read. In the original code that would always be the request size, regardless of the actual size of the data.
    That did not make sense to me at all.

5. However, the original value of tmp is req->td->dmacount, which is the dmacount value when the request's dma transaction was finished. And that is a much more reasonable value to report back to the caller.

As you can see, this is based a lot on educated guesses and debug printouts of various cases. That is why I would like to get your input on this, to make sure I'm on the right track.

To recreate the problem., try reading from a bulk out endpoint in a loop, 1024 * n bytes in each iteration. Connect the PLX to a host you can control, and send to that endpoint 1024 * n +x bytes such that  0 < x < 1024 * n and (x % 1024) != 0 You would expect the first read() to return 1024 * n and the second read to return x, but you will get the first read to return 1024 *n and the second one to return 1024 * n.
That is true for every positive integer n.

My patch solves the problem, and does not break any of the other cases I've tried.


getsadzeg:
It has conflicts with `drivers/usb/gadget/udc/net2280.c` , doesn't it?

PR #396, opened 2017-03-20
remark header.S

AnanSN:


AnanSN:
first merge

arch/x86/boot/header.S
```diff
@@ -12,7 +12,8 @@
  * addresses must be multiplied by 16 to obtain their respective linear
  * addresses. To avoid confusion, linear addresses are written using leading
  * hex while segment addresses are written as segment:offset.
- *
+ * 
+ * 在实模式使用64k段,所以,段地址 * 16 = 线性地址
```
songjiayang: don't use Chinese , change to English please.

daxtens:
There is a documentation section in Chinese - maybe that would be an
appropriate place for this contribution?
https://github.com/torvalds/linux/tree/master/Documentation/translations/zh_CN

arch/x86/boot/header.S
AnanSN: I will close it.

PR #398, opened 2017-03-29
asus-nb-wmi: Add wapf4 quirk for the X302UA

puruscor:
Asus laptop X302UA starts up with wifi hard-blocked without any meaningful way to enable it.
The same issue occurs on some other ASUS notebooks as well.

Setting wapf=4 fixes the problem.


ajdlinux:
Hey @puruscor, I'm not all that familiar with this particular subsystem but if you'd like a hand in figuring out how to get this upstream, let me know and I may be able to give you some hints.

PR #402, opened 2017-04-07
bots

gugu111:


LulzKiller:
Looks like spam.

daxtens:
Sadly the bot is not really equipped to detect spam. :(

PR #414, opened 2017-05-16
Tracking: allow CAP_SYS_RESOURCE override of quotas

kdave:
V2, https://marc.info/?l=linux-btrfs&m=149453743102171
"This patchset makes it so that on a per-filesystem basis one can disable
quota enforcement for users with cap_sys_resource. This patchset can
likely later be extended to per-qgroup, or a per-volume basis. I'm
thinking of extending the sysfs interface to list the qgroups and
this same interface for the qgroups themselves."

Target: 4.13


kdave:
Oh dang, sorry, wrong target tree.

PR #445, opened 2017-08-04
append Qualcomm GOBI 2K chipset ID for Panasonic CF-U1 Toughbook

Ole2mail:
just adding VID/PID for the item being verified as working Gobi 2K chipset

### DIFF
--- a/drivers/usb/serial/qcserial.c
+++ b/drivers/usb/serial/qcserial.c
@@ -76,6 +76,8 @@ static const struct usb_device_id id_table[] = {
 	{DEVICE_G1K(0x1bc7, 0x900e)},	/* Telit Gobi QDL device */
 
 	/* Gobi 2000 devices */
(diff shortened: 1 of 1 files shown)


drivers/usb/serial/qcserial.c
```diff
@@ -76,6 +76,8 @@ static const struct usb_device_id id_table[] = {
 	{DEVICE_G1K(0x1bc7, 0x900e)},	/* Telit Gobi QDL device */
 
 	/* Gobi 2000 devices */
+	{USB_DEVICE(0x04da, 0x250e)},   /* Panasonic Gobi 2000 QDL device */
+	{USB_DEVICE(0x04da, 0x250f)},   /* Panasonic Gobi 2000 Modem device */
```
wojtekmaj: You used spaces while all the other comments are separated from the code by tabs :)

Ole2mail:
hello Wojciech
thanks a lot.
I've pushed with tabs already.
I believe the next step is to cook the patch?
Oleg

PR #470, opened 2017-09-26
performance fix

parduscix:
change for -> while

### DIFF
--- a/init/main.c
+++ b/init/main.c
@@ -267,7 +267,9 @@ static int __init set_init_arg(char *param, char *val,
 
 	repair_env_string(param, val, unused, NULL);
 
-	for (i = 0; argv_init[i]; i++) {
+	i = 0;
+	while(argv_init[i]){
+		i++;
 		if (i == MAX_INIT_ARGS) {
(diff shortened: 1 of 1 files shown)


init/main.c
```diff
@@ -267,7 +267,9 @@ static int __init set_init_arg(char *param, char *val,
 
 	repair_env_string(param, val, unused, NULL);
 
-	for (i = 0; argv_init[i]; i++) {
+	i = 0;
+	while(argv_init[i]){
```
siyiwolf: why change while

klingtnet:
Now I know why the kernel development happens in mailing lists, this _performance fix_ is a joke, right?

MrNirva:
I agree with you

alexozer:
What the debate about, this seems pretty serious to me? It's a simple change for a simple improvement.

init/main.c
zwang001: Any special gain?

kalman5: Also on first iteration i = 1 not 0 as it is with the for

PR #471, opened 2017-09-26
Add author's name

Mandroide:
It looks like some lines on the LICENSE hasn't been added yet.

For instance: 332, 349, 309, 310.


arnavb:
:/ That isn't how the GNU License works....

PR #473, opened 2017-09-29
Update addrlabel.c

SoundButton:
Why the logical &&?
Would like to add this will lead to problems when p->ifindex = 0.


parrotgeek1:
You've managed to close this then reopen it. Please close it

PR #489, opened 2017-11-18
Suggestion for code style fix.

fmmedeiros:
Adding brackets to show explicitly which else belongs to which if statement.


cdown:
(just fyi, as is this is against kernel style guide, so that should be updated too)

PR #490, opened 2017-11-21
Remove gotos

christianbender:
I have remove some gotos in functions:

- cap_ptrace_access_check

- cap_ptrace_traceme

This is my first open source project. I would be pleased over feedback.


nuskorpios:
Your attemp of "improvement" just made the code almost unreadable. I mean, why would one want to chain *multiple* conditions with `&&` and `||`, repeat code and burst the line width limit just to avoid the use of "evil" `goto`s?

You should have checked the [Linux Kernel Style Guide][guide] first. And read some serious C code too.

[guide]: https://www.kernel.org/doc/html/latest/process/coding-style.html

christianbender:
@nuskorpios thanks for your feedback.

PR #493, opened 2017-11-27
x86/tsc: Add kernel parameter to override TSC frequency

Nekotekina:
Add kernel parameter tsc_freq= (in KHz) to override TSC frequency returned by native_calibrate_tsc()

### DIFF
--- a/arch/x86/kernel/tsc.c
+++ b/arch/x86/kernel/tsc.c
@@ -280,6 +280,16 @@ static int __init tsc_setup(char *str)
 
 __setup("tsc=", tsc_setup);
 
+static int tsc_freq_override_khz;
+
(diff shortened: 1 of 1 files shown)


arch/x86/kernel/tsc.c
```diff
@@ -280,6 +280,16 @@ static int __init tsc_setup(char *str)
 
 __setup("tsc=", tsc_setup);
 
+static int tsc_freq_override_khz;
+
+static int __init tsc_freq_setup(char *str)
+{
+	tsc_freq_override_khz = simple_strtol(str, NULL, 0);
+	return 1;
```
getrembl: Hi,
Why always `return 1`?
Couldn't you just replace the return type by `static void`?

PR #496, opened 2017-11-30
Add new QCA Rome Device (04ca:3015)

ioanmoldovan:


ioanmoldovan:
I already sent it via mailing list, [https://www.spinics.net/lists/linux-bluetooth/msg71679.html](https://www.spinics.net/lists/linux-bluetooth/msg71679.html). 

@holtmann Please take a look at it. Thank you!

PR #499, opened 2017-12-04
Added Dell 1537

guylamar2006:
Added Dell 1537 Vendor ID 0x19


ajdlinux:
Hi @guylamar2006,

Let me know if you need help with getting this patch upstream - the information in the bot message above should give you enough pointers to get started.

(ping @daxtens)

PR #503, opened 2018-01-04
x86: warn if unsecure CPU is being used

teknoraver:
Signed-off-by: Matteo Croce <mcroce@redhat.com>


aviraxp:
PS v2: Add missing \n

teknoraver:
never mind, it was a joke just to get a valid hash :P

PR #504, opened 2018-01-05
Update common.c

KrzysztofSzewczyk:
Update file because AMD actually IS affected by meltdown/spectre attack(s):
https://securityledger.com/2018/01/google-details-cpu-flaws-says-amd-arm-intel-affected/
https://tech.slashdot.org/story/18/01/04/0524234/google-says-almost-all-cpus-since-1995-vulnerable-to-meltdown-and-spectre-flaws
" [...] Spectre affects AMD and ARM as well (and likely other architectures too). 
Best I can tell, the only CPUs guaranteed not affected by both are in-order architectures, which many older ARM (and extremely old x86) chips are.
These attacks are a sort of new category of security analysis--realizing that out of order execution can have side effects, and that programs can check for those side effects to leak program state and system memory. "
So yes, it would be safer to just don't touch this modified code - at least we are sure that CPU will be protected in case of any attack.

### DIFF
--- a/arch/x86/kernel/cpu/common.c
+++ b/arch/x86/kernel/cpu/common.c
@@ -923,8 +923,7 @@ static void __init early_identify_cpu(struct cpuinfo_x86 *c)
 
 	setup_force_cpu_cap(X86_FEATURE_ALWAYS);
 
-	if (c->x86_vendor != X86_VENDOR_AMD)
-		setup_force_cpu_bug(X86_BUG_CPU_INSECURE);
+	setup_force_cpu_bug(X86_BUG_CPU_INSECURE);
 
 	fpu__init_system(c);
 


zc0de1rx:
This patch is effective only against Meltdown.

This change exists because AMD processors are not subject to the types of attacks that the kernel
page table isolation feature protects against. The AMD microarchitecture
does not allow memory references, including speculative references, that
access higher privileged data when running in a lesser privileged mode
when that access would result in a page fault.

KrzysztofSzewczyk:
What about Spectre?

zc0de1rx:
Meltdown uses a bug in the Intel and ARM processors, which, due to speculative execution of instructions, the processor ignores memory access rights.

Specte uses the feature of the work of prediction algorithms for branches and transitions in modern processors, because of which one process can affect the probability of speculative execution of instructions in another process.

Specte does not imply direct access to foreign memory in any form, even with speculative execution. Instead, Specte does so that the attacked process (this can be both the kernel of the system and another user program) itself gave out information about the contents of its own memory.

Imagine in the code of the attacked process such a construction, and the variable x is a consequence of some user input, to which we can influence:

if (x < array1_size)
{
    y = array2[array1[x]];
}

Now we take and write in our program exploiting the vulnerability, the most similar design, and we perform it many, many times, and each time the honestly calculated condition yields true, the array indices are completely valid, and in general everything is fine. The branch prediction unit thus collects statistics stating that this construction is always evaluated to true, so when you meet it, you can not wait for the calculation to finish but go directly to the content.

And now we transfer such data to the attacked program that x suddenly jumps out far beyond the array1 array. If there was no speculative execution, the processor would count the condition x < array1_size, find it invalid and jump over. But it is, and the prediction block tells him that x < array1_size will almost certainly be executed, so while somewhere from the memory slowly and sadly the value of array1_size is sucked to actually perform the comparison, the processor starts executing the body of this piece of code.

Since with Specte we are dealing with the fact that the processor learns to perform one process faster by the example of executing another process, thereby actually allowing the second process to monitor the progress of the first one, enabling this fix for the AMD will do nothing.

KrzysztofSzewczyk:
Closing.

PR #505, opened 2018-01-06
fix a bug refer to max_discard is 0

catShit:
typically TRIM cost shorter time than ERASE, so we expect max_trim as max_discard. But when ERASE timeout is too long, it did happened actually rather than theoretically.  A type of Toshiba eMMC's ERASE_TIMEOUT_MULT is too long and controller's max_busy_timeout is relatively small, then max_discard could be 0, even max_trim is available, but max_trim < max_disable can't be true, so we get a result of 0 instead of max_trim.


ajdlinux:
Not an expert in this driver, but if you need help with getting this upstream and the instructions above aren't enough let me know

aktiger87:
unsubscribe

catShit:
https://patchwork.kernel.org/patch/10816817/
https://patchwork.kernel.org/patch/10833425/

PR #507, opened 2018-01-15
experimental unblock

claymodel:
Check if the blocking and unblocking works well for creating an effect of 
```perpetual force```


claymodel:
Its experimental, and want to remove the PR as still work in process.

close #507

PR #509, opened 2018-01-22
Sorting error

kathryntolsen:
Matt Mackal was placed at the top out of order. Recommending a pull to make the document follow the proposed sorting rules Linus intended.


ciscoxll:
@torvalds  I love you.

PR #515, opened 2018-02-05
Updates calculation of wTotalLength for UAC2 gadget USB descriptor

robiwano:
According to specification (http://www.usb.org/developers/docs/devclass_docs/audio10.pdf, page 38), the descriptor header size should be included in wTotalLength. This **should** fix problem with UAC2 gadgets not getting recognized in Windows 10.


ajdlinux:
I'm not a USB expert by any means, but let me know if you need help with guiding this patch upstream per the info above.

robiwano:
Thank you, I've read through the submission instructions and will issue a patch within a week, I'll get back to you if I have queries. But, although the patch is a "no-brainer", I want to test it on my RPiZ first :)

ajdlinux:
@robiwano awesome :)

robiwano:
Unfortunately fixing wTotalLength did not help with the device not getting recognized :( , still is a bug though...

PR #519, opened 2018-02-15
USB: serial: cp210x: add new device EZP2010

klukonin:
EZP2010 is a very popular Chinese USB SPI Programmer. Manufacturer provides only windows drivers, but this device is based on Silicon Labs CP210x chip, so it is not so hard to add it to the driver id table.


ajdlinux:
Let me know if you need help upstreaming this, the instructions above should give you most of what you need to know.

EEpimetheus:
I'm brand new to linux trying to use my EZP2010 for the first time since moving from windows. I'm on Ubuntu 18, can somebody tell how to do this or point me in the right direction? Thank you.

klukonin:
@EEpimetheus Sorry, it was a kind of initial commit. But I didn't make a real support of this pprogrammer (I have one in my lab).
Last time my attempt was unsuccessful and I think the only way to move forward is to do some reverse engineering.
I't a little bit strange that such programmer "emulates" third-party ttyUSB interface.
May be we should move to another place and continue our work =)

EEpimetheus:
weird alright thanks. Alright, at least programmers are cheap.

PR #528, opened 2018-03-05
A question about your kernel 4.12.1

mickpf:
Hello,

is your kernel compatible/usable with UP Squared also?

Thanks in advance,
Michael


kittyhacker101:
After looking at your commits, although the kernel will probably work with most, if not all, UP Board attachments, they aren't going to merge it with the main kernel. The changes are likely to also work with the UP Squared, but due to differences in hardware, it is likely that not all functionality will work properly. I don't have an UP Squared (I have the normal UP Board), so I can't test this myself, but you could try it out and see.

kaisersGa:
l've seen it



在 2018-03-11 11:05:49，"Kat" <notifications@github.com> 写道：


After looking at your commits, although the kernel will probably work with most, if not all, UP Board attachments, they aren't going to merge it with the main kernel. The changes are likely to also work with the UP Squared, but due to differences in hardware, it is likely that not all functionality will work properly. I don't have an UP Squared (I have the normal UP Board), so I can't test this myself, but you could try it out and see.

PR #546, opened 2018-04-06
Removed the non working back-quotes from README.md  around make

ShivajiReddy:
For some reason _make htmldocs_ is not changing into `make htmldocs` in README.md even after trying in multiple ways. So just removed the back quotes which were serving no use.


Audition-CSBlock:
Try adding triple back-quotes, ```like this``` (```)

PR #554, opened 2018-05-28
ideapad-laptop: Add E42-80 to no_hw_rfkill list

dylanchu:
`Lenovo Zhaoyang E42-80` is another Lenovo ideapad model **without a hw rfkill switch**, resulting in wifi always reported as hard blocked.
Add it to the list of models without rfkill switch.


dylanchu:
The patch has been sent following the instructions and I'm now waiting for the reply.

PR #574, opened 2018-07-30
ath9k_htc: increase number of configurable virtual interfaces

psyborg55:
Testing with AR9271 USB device I was able to start 8 VAPs, but one client device could connect only if 7 VAPs were configured. This might need more work to make the feature fully usable, but even at present situation it enables functionality that was previously disabled.
Updated ath9k_htc firmware is required too, PR already opened: https://github.com/qca/open-ath9k-htc-firmware/pull/149

Signed-off-by: Tomislav Požega <pozega.tomislav@gmail.com>

### DIFF
--- a/drivers/net/wireless/ath/ath9k/htc.h
+++ b/drivers/net/wireless/ath/ath9k/htc.h
@@ -197,8 +197,8 @@ struct ath9k_htc_target_rx_stats {
 	__be32 host_done;
 } __packed;
 
-#define ATH9K_HTC_MAX_VIF 2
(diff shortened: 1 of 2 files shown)


ajdlinux:
Hi @psyborg55 - are you planning to submit these patches to the upstream kernel?

psyborg55:
Hi

I thought this is the upstream kernel repo? Where the hell did i end up? On microsoft kernel?

ajdlinux:
I would suggest reading the message from the bot at the top of this thread :) Let me know if you have questions.

PR #576, opened 2018-08-06
tlb_uv.c: Add missing return value check to prevent nptr dereference.

Jason2031:
Function init_per_cpu() defined in arch/x86/platform/uv/tlb_uv.c calls kzalloc() to allocate memory for variable uvhub_mask which is used in function get_cpu_topology() and function summarize_uvhub_sockets() later. However, none of the 3 functions mentioned above checks whether uvhub_mask is NULL. As kzalloc() may return NULL, when get_cpu_topology() and summarize_uvhub_sockets() tries to dereference this pointer, it may cause NULL pointer dereference bug.


Jason2031:
Fix bugzilla report [200489](https://bugzilla.kernel.org/show_bug.cgi?id=200489).

PR #577, opened 2018-08-06
pktcdvd.c: Fix wrong return code when alloc_disk() fails.

Jason2031:
Function pkt_setup_dev() defined in drivers/block/pktcdvd.c calls alloc_disk(). However, it forgets to set the error return code when alloc_disk() fails. Instead, when alloc_disk() fails, it simply jumps to label 'out_mem' leaving the variable ret unchanged.


Jason2031:
Fix bugzilla report [200505](https://bugzilla.kernel.org/show_bug.cgi?id=200505).

PR #578, opened 2018-08-07
conntrack.c: Add missing return value check to prevent nptr dereference.

Jason2031:
Function ovs_ct_limit_cmd_get() defined in net/openvswitch/conntrack.c may cause a null pointer dereference as it calls nla_nest_start which may return NULL. The returned value is used in function nla_nest_end() later where the pointer is dereferenced.


Jason2031:
Fix bugzilla report [200535](https://bugzilla.kernel.org/show_bug.cgi?id=200535).

PR #591, opened 2018-09-17
I think "code-of-conflict" and "code-of-product" should be removed because they will incite hatred

protonesso:


protonesso:
I think this code will suppress free speech and will incite hatred. Let's live in free world :smile:

PR #597, opened 2018-09-21
Code cleanup in linux/sched/

PierceGriffiths:


PierceGriffiths:
You should bring that to the attention of someone who actually cares, @benixlinux

bmeh:
> The goal of this issue is not to stir up arguments

Yet he spams entire GitHub with this nonsense. How is this relevant to this pull request?

kernel/sched/cpupri.c
```diff
@@ -29,20 +29,16 @@
...
 
 /* Convert between a 140 based task->prio, and our 102 based cpupri */
-static int convert_prio(int prio)
+static int convert_prio(const int prio)
 {
-	int cpupri;
-
 	if (prio == CPUPRI_INVALID)
-		cpupri = CPUPRI_INVALID;
+		return CPUPRI_INVALID;
```
gxgalx: Isn't it clearer having just one return statement instead of several exit points?

kernel/sched/rt.c
```diff
@@ -142,10 +142,12 @@ void free_rt_sched_group(struct task_group *tg)
 		destroy_rt_bandwidth(&tg->rt_bandwidth);
 
 	for_each_possible_cpu(i) {
-		if (tg->rt_rq)
-			kfree(tg->rt_rq[i]);
-		if (tg->rt_se)
-			kfree(tg->rt_se[i]);
+		/* Don't need to check if tg->rt_rq[i]
```
gxgalx: It was actually checking if tg->rt_rq and tg->rt_se are NULL or not, not the elements in the array.

PierceGriffiths: I don't know how I managed to overlook that. I've received feedback from Peter Zijlstra describing this patch as "horrible" and I agree with that assessment. I'm waiting for a response from Peter regarding whether he thinks any part of the patch is salvageable or if the whole thing is better off being thrown out.

asm-pas:
I don't think your code follows the "linux-code-style"

kernel/sched/cpupri.c
GTANAdam: It's a tiny optimization that should reduce the amount of instructions used in that method imo (if the compiler doesn't optimize that already), instead of assigning cpupri variable a value and then return it, it will just return the value instead.

PR #594, opened 2018-09-19
Let's re-insert the Code of Conflict. Greg Kroah-Hartman is an usurper.

jason-oliveira:
Undoing the Freedesktop.org coup. Linus was blackmailed.

PR #599, opened 2018-09-26
Replaced the Code of Conduct with the Code of Conflict

AshtonSnapp:
Due to the amount of controversy following the installation of the Code of Conduct, I am opening this pull request to replace it with the original Code of Conflict, to, as put by Joey Pabalinas, "Unfuck our Codes of Conduct", and to stop the controversy and drama (two things I never liked).

### DIFF
--- a/Documentation/process/code-of-conduct.rst
+++ b/Documentation/process/code-of-conduct.rst
@@ -26,12 +26,3 @@ As a reviewer of code, please strive to keep things civil and focused on
 the technical issues involved.  We are all humans, and frustrations can
 be high on both sides of the process.  Try to keep in mind the immortal
 words of Bill and Ted, "Be excellent to each other."
-
-A Note from Me (Ashton)
------------------------
-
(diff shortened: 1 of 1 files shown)


SteveHeist:
I'd vote in favor of #594 - does the same thing

asm-pas:
@SteveHeist I would vote for anything that gets us back on track

ShadowGoat:
Yes do this now

SilicaAndPina:
i agree, should merge.

JMP2DIS:
I agree.
We have many people committing from different backgrounds all over the world, because of this people hold their own ideals. This could prevent a patch being made just because the committer holds ideals that conflicts with the CoC. Not to mention some people that make commits might not have a lot of social experience or may interpret things differently than others. Some people have a hard time distinguishing right from wrong, but should we be exiling or denying them just because we think something is right? If you want to be inclusive you want free flow of thought. The CoC imposes that gender, sex, race is relevant to committing to a open source project, but this counterproductive. If we just ignore the fact that we are different then people will treat each other the same not the other way around because by identifying with a group you create exclusiveness.

felipemullen:
@SilicaAndPina thanks for sharing that, it really opened my eyes...

Have the rest of you read the things CoralineAda say?? I'm seriously at a loss for words. It's just pure hate and malice. 

Is this what we are supposed to stand by? Because I won't. I choose to respect ALL people. Not a chosen inside group that aligns with my own world views.

PR #626, opened 2018-11-27
Add E-mu 0404 USB automatic sample rate

khseal:
Add old work patch...

### DIFF
--- a/sound/usb/quirks.c
+++ b/sound/usb/quirks.c
@@ -1114,7 +1114,8 @@ static void set_format_emu_quirk(struct snd_usb_substream *subs,
 	 * by playback substream
 	 */
 	if (subs->direction == SNDRV_PCM_STREAM_PLAYBACK) {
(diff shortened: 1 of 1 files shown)


khseal:
Temporarily close the request. It seems in the latest kernels auto-select works correctly.

sound/usb/quirks.c
```diff
@@ -1114,7 +1114,8 @@ static void set_format_emu_quirk(struct snd_usb_substream *subs,
 	 * by playback substream
 	 */
 	if (subs->direction == SNDRV_PCM_STREAM_PLAYBACK) {
-		if (subs->stream->substream[SNDRV_PCM_STREAM_CAPTURE].interface != -1)
+		if (subs->stream->substream[SNDRV_PCM_STREAM_CAPTURE].interface != -1 &&
```
AlaaAttya: i think it would be better if you could check the index of the array before access it (save array access)
so you need to check if `subs->stream->substream[SNDRV_PCM_STREAM_CAPTURE]` is not `NULL`

PR #638, opened 2019-01-03
Fixes Typo when Spind and wait around

Brakoman:


atakanyenel:
This is not a typo. `While` command executes the first line under it if there are no `{` defined.

PR #649, opened 2019-02-07
Fixed the README Format.

josephworks:
It was not rendering as markdown because it was not a .md file.
I also added links for mentioned documents.


josephworks:
I am not creating a patch, I am just updating the readme... What do I do?

PR #652, opened 2019-02-15
Update and rename README to README.rst

khode-mohsen:


MilesBHuff:
On the one hand, I like this idea, because it would make GitHub display the ReadMe a little more prettily.  On the other hand, "README" with no extension is traditionally correct.

PR #655, opened 2019-02-23
[vimrc]: cscope: implement cnext/cprev shortcuts

LazarRazvan:
Course presents F5/F6 keys as shortcuts to move between multiple
results but they are not implemented (not working).

"If there are more than one results (usually there are) you can
move between them using F6 and F5"
[cscope_shortcuts.zip](https://github.com/torvalds/linux/files/2896903/cscope_shortcuts.zip)


techonsa:
Hi 

My apologies if commenting here is incorrect and if so please direct me to the correct place to ask about this problem. My Raspberry Pi 3 wireless is not working properly after upgrade to Buster and rpi-update to latest firmware. Is it a kernel modules issue? Where does the rpi-update command download from?

PR #661, opened 2019-03-08
Fixed typo: extra space.

jacob-macleod:
There was a double space.


Josh1Dev:
@jacob-and-nathan Did you really create a whole pull request for a double spaced comment? Please...

dicom-nate:
He can now put "contributed to Linux kernel" on his resume.

ssbarnea:
I have seen tons of contributions like this at OpenStack and even worse: ones generated with bots, where you can see that the same account fixes typos in 30 repositories during few minutes.

Most of the time those fixes were valid, but was really annoying was that I seen different fixes proposed for each typo. That was fun, not.

On the other hand, lets try not to be too negative about someone's first attempt to contribute. In the end it was a valid fix, the mistake was only the place where it was raised.

PR #662, opened 2019-03-13
gpio-f7188x: add support  for Fintek F81965

Smiche:
As per specifications http://www.hardwaresecrets.com/datasheets/F81865_V028P.pdf
Updated logs for supported devices added from previous commit


Smiche:
my bad

PR #669, opened 2019-04-06
Fixed spelling error in docs

jakegore:
Fixed errors on the memory management page that were showing on the kernel docs


ilMagnifico:
The time you used to look up grammatical errors in all the docs, you could have used it for something more profitable.

jakegore:
I was just reading the docs and came across it. I thought I may as well take a couple of seconds to change it.

ericm:
This is a readonly mirror of the kernel if you're not aware.

jakegore:
Ah right, thanks. I haven't done anything like this before 👍 Cheers anyway

PR #674, opened 2019-05-03
Ideapad S205 has a hardware switch but den ACPI driver reads the hard…

JohnCramer66:
…ware state wrong, so the wifi is hard blocked and not working.

Instead of adding hundreds of exceptions of lenovo laptops, why not adding a switch for loading the module with wifi hardware state unblocked or something like this,


XuZhongCN:
The way to read "hardware rfkill switch" is not necessarily the same on every laptop.
Therefore, judging the type of laptop cannot be omitted.

PR #687, opened 2019-06-05
Fixed some typo and newline issues.

shaswata56:
some simple fixing and making more efficient.

### DIFF
--- a/init/initramfs.c
+++ b/init/initramfs.c
@@ -30,7 +30,6 @@ static ssize_t __init xwrite(int fd, const char *p, size_t count)
 		out += rv;
 		count -= rv;
 	}
-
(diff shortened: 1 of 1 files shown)


hbina:
Perhaps I am misguided, how is doing the same thing twice more performant?

shaswata56:
Thanks for your reply. I was thinking about memory. Isn't kernel memory so precious?

hbina:
No, this is not a dynamically allocated memory so it will go out scope as soon as this function is done. Additionally, a single integer worth nothing compared to having to evaluate the same values twice.

shaswata56:
Thanks for your explanation. :+1:

PR #688, opened 2019-06-08
Removed goto and used return.

shaswata56:
Results are same. But, more convenient.


hbina:
Hi, this is a perfectly fine commit. Please consider following the instruction on how to make contributions.

> Linux kernel development happens on mailing lists, rather than on GitHub - this GitHub repository is a read-only mirror that isn't used for accepting contributions. So that your change can become part of Linux, please email it to us as a patch.

It is relatively easy to follow : https://kernelnewbies.org/FirstKernelPatch

shaswata56:
Thank you. I'll try to send the patch asap..

PR #694, opened 2019-06-26
Significant kernel performance improvements

lucavinci:
I made huge improvements to the linux kernel by optimizing the code to not even compile at all so the runtime will be very close to O(0). This should make the kernel more usable for most users.


TwistedPrime:
Very pru

shagbag913:
Can confirm 10x performance improvement

maxh65000:
Also much more power efficient.

person4268:
As much as this increases performance, it sadly removes compatibility for many platforms.

PR #698, opened 2019-07-05
Remove the extra space characters at the end of lines for each file

xianfengting:
# Remove the extra space characters at the end of lines for each file in Linux kernel source  
Because my email box can't send patches to LKML, so I created this PR.  
I will also use my friends' email box to send my patch as soon as I can.  
These changes may have something wrong, I will also have a check and improve, so please have a check before merging this PR.


ImPanda123:
just WHY XD ?

PR #702, opened 2019-07-14
Updated and renamed Update to Update.md

MrMe42:
Moved to markup file, added hyperlinks, added welcome, added thanks.


SamHannat:
hmmm this is cool but not cool enough

PR #705, opened 2019-07-31
Improved doc-comment in stackleak_plugin.c

deeprobin:
Cleaned documentation comment up. I removed the "TODO" because it was very old.


ajdlinux:
Hi @DeepRobin - thanks for looking at this. To remove a TODO comment like this, you'll have to justify it and explain how you've shown that the existing cases do actually cover everything, not just that it's an old comment.

PR #710, opened 2019-08-14
Update security.c

kafirhamsaeiyya:


security/security.c
```diff
@@ -1,2360 +1,6 @@
...
-int security_kernel_module_request(char *kmod_name)
-{
-	int ret;
-
-	ret = call_int_hook(kernel_module_request, 0, kmod_name);
-	if (ret)
-		return ret;
-	return integrity_kernel_module_request(kmod_name);
+#include <stdio.h>
+void main() {
```
JL2210: `void main()` is bad practice. Use `int main(void)` instead.

security/security.c
```diff
@@ -1,2360 +1,6 @@
...
+	printf("don't add this pull request")
```
JL2210: Instead of using `printf` here, use `puts`.

JKJameson:
Do you work for Intel?

security/security.c
kafirhamsaeiyya: wth lol hahaha

security/security.c
kafirhamsaeiyya: um

wangxing1986: very good.
The puts function writes string to the standard output stream stdout, replacing the string's terminating null character ('\0') with a newline character ('\n') in the output stream.

PR #713, opened 2019-08-21
Update README,added languages

ravachol-yang:
I've made a better README file;
It looks beautiful;
And I've added some more languages,I think it will help many foreign people **understand the README better**

Please accept my change ,it will help many people 
_**Thanks a lot!!!!!**_

### DIFF
--- a/README
+++ b/README
@@ -1,18 +1,29 @@
-Linux kernel
-============
+                                                        ***Linux kernel***
+                                                            Linux   核心
+#English:_________________________________________________________________________________________________________________________________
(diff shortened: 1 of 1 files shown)


xwolfi:
You seem to only have modified an 11-line readme, added two empty variants of chinese with no translation and asked to be merged.

Could you instead:
- Create one README file per language, for all readmes in the kernel
- Insist on getting more language than chinese 
- Publish them all outside of the kernel, possibly on a wiki, so that people can actually use them.

PR #732, opened 2019-10-04
dtc: Use pkg-config to locate libyaml

pmod:
Makefile's wildcard use predefined location of yaml.h
header file which may lead to false-positive detection
of presence of libyaml installation when cross-compiling
(e.g. in yocto environment).
Use instead pkg-config to detect libyaml.


pmod:
@robherring @masahir0y please review

pmod:
done

PR #740, opened 2019-10-09
Updating deprecated SPDX license headers

Frosticles:
In the 3.0 release of the SPDX license list the use of just `GPL-x.x` or `GPL-x.x+` was deprecated in favor of specifying `GPL-x.x-only` or `GPL-x.x-or-later`. 

I wrote a script to go through and update the outdated headers:
```
find . -type f -print0 | xargs -0 -I % perl -pi -e \
 's/(SPDX-License-Identifier:.*GPL-\d\.\d(?!(-only|-or-later|\+)))/$1-only/g' %

find . -type f -print0 | xargs -0 -I % perl -pi -e \
 's/(SPDX-License-Identifier:.*GPL-\d\.\d)(\+)/$1-or-later/g' %
```

The first regex statement is to replace `GPL-x.x` with `GPL-x.x-only`. It looks for `SPDX-License-Identifier:` followed by 0 or more of any characters except new lines, followed by `GPL-{digit}.{digit}`, and then a negative lookahead to discard the match if it is followed by `-only` or `-or-later` to ignore headers that have already been updated. The matched statement is appended with `-only` by taking the capture group 1 followed by -only with `$1-only`.

The second regex statement is to replace `GPL-x.x+` with `GPL-x.x-or-later`. It looks for `SPDX-License-Identifier:` followed by 0 or more of any characters except new lines, followed by `GPL-{digit}.{digit}+`. The statement is split into 2 capture groups so that the `+` character can be thrown away before appending `-or-later` with  `$1-or-later`.


Frosticles:
Doesn't look like a practical change

PR #764, opened 2019-12-16
Update panic.c

Error-Undefined:
We should really use modern terminology, in one month it's 2020

### DIFF
--- a/kernel/panic.c
+++ b/kernel/panic.c
@@ -212,7 +212,7 @@ void panic(const char *fmt, ...)
 	if (len && buf[len - 1] == '\n')
 		buf[len - 1] = '\0';
 
-	pr_emerg("Kernel panic - not syncing: %s\n", buf);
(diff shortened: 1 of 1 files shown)


JamesTheHacker:
Lame. Next ...

PF94:
That's a obvious yikes for me, This isn't a terminology, but in fact, a meme that will die in a few weeks.

kernel/panic.c
```diff
@@ -212,7 +212,7 @@ void panic(const char *fmt, ...)
 	if (len && buf[len - 1] == '\n')
 		buf[len - 1] = '\0';
 
-	pr_emerg("Kernel panic - not syncing: %s\n", buf);
+	pr_emerg("Vibe check - not syncing: %s\n", buf);
```
PF94: Not funny. Move on.

bazhenovrussia: What???

MachineThing: The kernel has not passed the vibe check

PR #771, opened 2019-12-30
Fix POSIX: POSIX too slow, Complete performance Overhaul

jobs-git:
http://www.cs.columbia.edu/~vatlidak/resources/POSIXmagazine.pdf

https://news.ycombinator.com/item?id=13621623

### DIFF
--- a/README
+++ b/README
@@ -1,6 +1,8 @@
 Linux kernel
 ============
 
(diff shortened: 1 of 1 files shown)


craigbarnes:
@jobs-git So where's your fix? Let us know when it's ready to merge.

PR #772, opened 2019-12-30
Fix alignment of value in /proc/$pid/smaps

ioquatix:
The `/proc/$pid/smaps` output has an alignment issue for the field `FilePmdMapped` and `THPeligible`.

This increases the alignment of `FilePmdMapped` by 1 space, and converts the alignment of `THPeligible` to use spaces instead of tabs, to be consistent with the other fields.


ioquatix:
Just for completeness/quick reference/review here is a picture of the alignment issues:

image

ioquatix:
This was submitted to the fsdevel mailing list.

https://marc.info/?l=linux-fsdevel&m=157769534719460&w=2

PR #775, opened 2020-01-01
missing parentheses in USE_NEW_SCHEME

atmgnd:
accroding to https://github.com/torvalds/linux/commit/bd0e6c9614b95352eb31d0207df16dc156c527fa#diff-28615d62e1250eadc353d804f49bc6d6, will try old enumeration scheme first for high speed devices. for example, when a high speed device pluged in, line 2720 should expand to 0 at the first time. USE_NEW_SCHEME(0, 0 || 0 || 1) === 0. but it wrongly expand to 1(alway expand to 1 for high speed device), and change USE_NEW_SCHEME to USE_NEW_SCHEME((i) % 2 == (int)(scheme)) may be better ?

### DIFF
--- a/drivers/usb/core/hub.c
+++ b/drivers/usb/core/hub.c
@@ -2692,7 +2692,7 @@ static unsigned hub_is_wusb(struct usb_hub *hub)
 #define SET_ADDRESS_TRIES	2
 #define GET_DESCRIPTOR_TRIES	2
 #define SET_CONFIG_TRIES	(2 * (use_both_schemes + 1))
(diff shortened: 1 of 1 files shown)


atmgnd:
the missing parentheses causes a fail wiht "device descriptor read/64, error -110" using my usb driver on vmware using usb 3.0 hub.

atmgnd:
the missing parentheses causes a fail wiht "device descriptor read/64, error -110" using my usb driver on vmware using usb 3.0 hub.

PR #783, opened 2020-02-07
Delete MIT

jeepdude1980:


aaronchall:
I don't think this PR is submitted in good faith. See also: https://twitter.com/Nick_Craver/status/1225958276224946176

ryahpalma:
Why do you want to delete all licenses?

PR #841, opened 2024-04-14
pr_debug for vhci_hcd unlink->seqnum and urb->status

frostworx:
While using usbip as client, the journal is flooded with

```
Apr 14 12:12:28 mini kernel: vhci_hcd: unlink->seqnum 158446
Apr 14 12:12:28 mini kernel: vhci_hcd: urb->status -131
```
messages, because `pr_info` is used.

With this PR, I'd like to suggest using `pr_debug` instead,
because the info most likely is only interesting for debugging.


dukie:
stop

frostworx:
pardon?

ricky6672 (approved):

ricky6672 (changes requested):
- - Kslle_@_

frostworx:
of course the patch is not significant, but still, it would be nice to have it upstream _(especially on weak systems)_

PR #842, opened 2024-04-20
Renamed README to README.md for better readability

Mousemaster12:
Renamed README to README.md for better readability. made it more accessible


franticboDev (changes requested):
just why? Let it be, it doesn't hurt anybody.

Mousemaster12:
Its already written as a md file, changing it will just provide users with better readability in Github's format, and will require no work besides changing the file type

Magnusretrotech:
I was about to add this before I saw you had already done this. This is a very good idea.

orxngc (approved):

zhang-xue-feng:
"Why are there 3 commits needed to do this?"

Mousemaster12:
> "Why are there 3 commits needed to do this?"

it was a mistake, one is enough, just change it to readme.md. I accidentally  deleted the file, re added it and then only renamed it if my memory serves me right

Mousemaster12 (commented):
looks great

zhang-xue-feng:
> > "Why are there 3 commits needed to do this?"
> 
> it was a mistake, one is enough, just change it to readme.md. I accidentally deleted the file, re added it and then only renamed it if my memory serves me right

i think a squashing rebase can be made in this case

LiteApplication:
This could break a lot of websites linking to Linux's README, without providing that much of an increase in clarity.

Also maybe squash your commits into one and rebase.

PR #851, opened 2024-04-29
Tiny update mem.c

fantik86:


aiekick:
why this change ????

aleao-sl:
> why this change ????
Good question... An if block with only one instruction and removed enclosing brackets. It looks like an irrelevant update. does not add benefits in maintainability

fantik86:
> > > why this change ????
> > > Good question... An if block with only one instruction and removed enclosing brackets. It looks like an irrelevant update. does not add benefits in maintainability, but not a bad idea
> 
> i aggree, firt time im comming on this repo, and im so surprised to see many idiot PRs like this one. strange....

I am very sorry, this is my very first pull request...

Bug-Reaper:
@aiekick @fantik86 this repo is a read only mirror of linux. No PRs are ever accepted here.

PR #853, opened 2024-04-30
Change watchdog to watchcat

kur00n3k0:
Changing the name watchdog to watchcat cuz cats are better than dogs!


Elsa-is-My-Muse:
66666

> Changing the name watchdog to watchcat cuz cats are better than dogs!将“看门狗”名称更改为“看门猫”，因为猫比狗更好！

What an interesting idea

miiiiiYT:
I'm nowhere near an expert on the kernel, however, despite this being a funny idea which a part of me likes to see in the actual kernel, this will probably have an immense effect on other bits of the kernel or just programs in the entire linux universe which rely on watchdog being named watchdog.   
I'd love to see this, however, the effort that's most likely tied to this change, sadly won't be worth the joke. :c

aleao-sl:
> I'm nowhere near an expert on the kernel, however, despite this being a funny idea which a part of me likes to see in the actual kernel, this will probably have an immense effect on other bits of the kernel or just programs in the entire linux universe which rely on watchdog being named watchdog. I'd love to see this, however, the effort that's most likely tied to this change, sadly won't be worth the joke. :c

Agree, just reading the name, makes me think it's a new feature instead of just renaming stuff. And cats are not guard animals.

tigercoding56:
> .... And cats are not guard animals.
how about watchtiger  then no process will mess around

dlimon2:
watchoctocat

haja-fgabriel:
image

PR #856, opened 2024-05-03
Fixes vmware failure to build modules on Fedora 40

dr3mro:
on Fedora 40 and kernel 6.8.8-300.fc40.x86_64 VMWare failes to build modules,  complaining of missing function

this is a fix for this issue that results in vmmon and vmnet kernel modules fails to build I hope its useful
Cheers


dr3mro (commented):
reoeder line

sheep4499:
why？timex.h

PR #858, opened 2024-05-06
convince microsoft to stop wsl

YeesterPlus:
You don't ever need windows,but instead, LINUX!

### DIFF
--- a/README
+++ b/README
@@ -16,3 +16,5 @@ several of them using the reStructuredText markup notation.
 Please read the Documentation/process/changes.rst file, as it contains the
 requirements for building and running the kernel, and information about
 the problems which may result by upgrading your kernel.
+
+Please do not let WSL go any further, MICROSOFT!


veberdev (approved):

ArchUser123:
+1

Odex64:
Bruh

sword-2:
Would version control between the MAINTAINER file in the source for both projects, where it showed the maintainers were the same, reduce concern about the WSL?

This PR may be supported by a VA page which identifies the WSL technology is not portable as it runs only on Windows platforms. That page is here https://www.oit.va.gov/Services/TRM/ToolPage.aspx?tid=14422&tab=2&minYear=2025#

A spot check of the MAINTAINERS file in torvalds/linux and WSL2 (https://github.com/microsoft/WSL2-Linux-Kernel) found the first 3 maintainer contacts were the same.

AvazRakhimov:
we need wsl, keep improving it!

ArchUser123:
make wsl have bugs so people switch to real linux

PR #860, opened 2024-05-08
replaced readme with the text MOTHER Tatyana Kurtukova

gmehdiev:
Added braille font, I think it is extremely inappropriate that Linux does not care about people with low vision


Cody212:
Bruh😭

p-shcherbatykh:
crazy

OTonGitHub:
Excuse my ignorance but, how will they actually read it though? I don't think blind people print repos into braille paper.

ASHWIN990:
🤣

a-usr:
Sorry, but I dont **see** how people would have an easier time recognizing a **pattern of tiny dots** compared to regular letters.

samwareing:
Visually impaired folks can, will and should use their screen reader (narration/braille tool) of choice.

This PR is, in my opinion, very ignorant.

Visually impaired people have better solutions to this universal problem, and this particular solution doesn't actually work.

nikitaspiridonovjuriba (approved):

OTonGitHub:
Ain't no way

a-usr (changes requested):
You are missing a `⠉` somewhere

ApplePieCodes:
> Visually impaired folks can, will and should use their screen reader (narration/braille tool) of choice.
> 
> This PR is, in my opinion, very ignorant.
> 
> Visually impaired people have better solutions to this universal problem, and this particular solution doesn't actually work.

ITS A JOKE

yusuftogtay:
lol

PR #861, opened 2024-05-09
French translations added

WedoneOfficiel:
French translations added


tools/power/cpupower/po/fr.po
0xNath, line 1: Hello,
I suggest `"Coeur Processeur XX"` instead of `msgstr "Processeur Core XX"` to stay coherent.

0xNath (approved):

ThisCatLikesCrypto:
Please sign this with your private key to prove identity. https://github.com/torvalds/linux/pull/17#issuecomment-5659933

PR #874, opened 2024-05-16
Update main.c

MorphyKutay:
The updated xbc_show_value function uses a recursive approach to print all the data in a boot configuration file. This function checks if a xbc_node is a value node. If it is, it prints the value to the screen. If not, it retrieves the child nodes of the current node and recursively calls the xbc_show_value function for each one. This allows all child nodes to be included in the output. This recursive approach effectively handles complex structures and nested data in a boot configuration file.


MorphyKutay:
++

admsev:
@gooroodev please review

AkshDesai04 (approved):

PR #893, opened 2024-06-15
Major Sanitizer is AndongSoju, which is tested

cholmaster:


cholmaster:
This can sanitize subsystem's data. Currently, my subsystem is sdns-rpl. See my repository.

cholmaster:
This is size-based sort buffer. This utilizes every incoming data as sorted format. You can code this into personal pathway to reset incoming packets. So example is suggested, but I don't think that this is general. So I close this issue.

PR #909, opened 2024-07-09
Update procfs-i8k

itzmew:
The abbreviation etc. always ends with a period, regardless of any additional punctuation that may follow. A comma precedes the use of “etc.,” whether abbreviated or written as “et cetera,” because it’s another item on the list.


VitalikObject:
what a sigma

muffinman987:
What a G find that's hilarious

PR #919, opened 2024-07-22
bpf: fix bpf_probe_read documentation

sonntex:
Fix bpf_probe_read documentation. Previously, the description might confuse because of the statement that bpf_probe_read function reads data from kernel space address only. In fact it compares passing address and TASK_SIZE, and then delegates execution to bpf_probe_read_user or bpf_probe_read_kernel functions.


chitturisuryaanuteja (approved):
LGTM

PR #925, opened 2024-08-03
Remove unwanted architecture from arch/ Directory

LinuxBoy-96:
This pull request removes support for all unwanted architectures from the arch/ directory, except for the following:

- x86
- ARM64
- RISC-V
- UML (User-Mode Linux)

The removal of these  architectures from the arch/ directory is only the first step to remove completely anything related to them in the kernel.
Drivers and Documentation will follow,


RelativisticMechanic:
lol

PewPewCricket:
This is the most retarded thing I've ever seen on the linux github mirror

PR #930, opened 2024-08-08
gpio: davinci: shortlog here

khilman:
a much more detailed description.

would go here.

Followed by even more details, and

Signed-off-by: Kevin Hilman <khilman@baylibre.com>
---
Then some trailers here, with diffstat and random stuff, and maybe some custom structured data, like

change-id: abcdefg
JIRA ID: [BLF-127]
prerequisite-message-id: foo-bar


khilman:
Closing dummy experiment

PR #938, opened 2024-08-23
Update code

didor-real:
.


jmarcher:
I always find it funny how many people find time to start a PR here and it is just utterly the worst changes ever

PR #944, opened 2024-08-28
Update uvc_driver.c

amilsanan:
add "ID 0408:4033 Quanta Computer, Inc. ACER HD User Facing" in uvc


jmarcher:
I always find it funny how many people find time to start a PR here and it is just utterly the worst changes ever

PR #947, opened 2024-08-30
reboot.c : changed reboot_default to bool

shubhamessier:
Made three changes:
1. ARM64 now defaults to hard reboot on kernel panic.
2. `reboot_default` flag is now a boolean for better clarity.
3.  comment changed to skip DMI Checking, code bypasses DMI quirk lookup if reboot is used.


Teyllayka:
exactly

MrZodoxVpython (commented):
yes

PR #948, opened 2024-09-04
Update crash_reserve.c: Optimizing Crash Kernel Memory Reservation, Fixing a few typos

LunerLA2:
- Typo Fix: Corrected the typo in the error message for the @ character check.

- Consistent Error Messages: Added more descriptive and consistent error messages.

- Dead Loop Prevention: Added a retry counter with a maximum limit to prevent infinite loops during memory reservation retries.

- Resource Management: Ensured that resources are cleaned up properly on error paths to avoid memory leaks and other issues.

### DIFF
--- a/kernel/crash_reserve.c
+++ b/kernel/crash_reserve.c
@@ -86,7 +86,7 @@ static int __init parse_crashkernel_mem(char *cmdline,
 		}
 		cur++;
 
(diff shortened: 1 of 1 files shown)


MonnXI (approved):
Just checked everything looks fine

PR #954, opened 2024-09-14
Update README

Glivermore2021:


Bashamega (commented):
I don't understand the change. It is quite useless.

amixsty:
I think this is a kind of spam so that he can register his name in the list of kernel contributors.

ruidazeng (commented):
What was the change?

PR #843, opened 2024-04-20
Fixes for Quanta Acer HD User Facing Web camera(original code from @Giuliano69)

Emtyffx:
There is a problem that I and some other users encounted while not being able to turn on camera on linux.
After searching for some time I have found fix created by @Giuliano69, thanking him a lot for this fix.

It will be good to merge this code into official uvc driver for not so experienced users, because they can't find the fix themselves(i also have spent some time finding the fix and applying it). And also the fix that require changing some kernel code can scare newbies.

PR #959, opened 2024-09-21
Fix for ID 0408:4033 Quanta ACER HD User Facing Camera

takshakmudgal:
This issue has mainly arises in the newer Acer Aspire 3 Series Laptops, where the cameras with this specific `ID 0408:4033` were not working on linux. 

After doing some extensive research a fix was found @Giuliano69 which solved the issue. This PR aims to implement that fix in the main kernel so that this issue gets fixed for all.
This issue was also referenced in this PR #843 but was left unattended and wasn't merged into the kernel, specifically the fix for ID 0408:4033, whereas ID 0408:4035 was merged.
Let's get this merged too, so that the issue is fixed for all users who are facing it.

### DIFF
--- a/drivers/media/usb/uvc/uvc_driver.c
+++ b/drivers/media/usb/uvc/uvc_driver.c
@@ -2430,6 +2430,24 @@ static const struct uvc_device_info uvc_quirk_force_y8 = {
  * though they are compliant.
  */
 static const struct usb_device_id uvc_ids[] = {
+	/**
+	   * Fix for the problem with cameras on Acer Nitro 5 Series & Acer Aspire 3 Series.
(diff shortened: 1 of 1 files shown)


MrZodoxVpython (commented):
gorgeous move

takshakmudgal:
> gorgeous move

send this through the mailing list aswell, hope to get it patched soon

Ten0:
Likely this should not contain all the extra "fix for blabla" comments, like for every other device below.

takshakmudgal:
Yes, I have talked with a maintainer and they said it'll probably be merged
with the 6.13 Kernel.

PR #960, opened 2024-09-22
perf evsel: display dmesg command of showing a hardcoded path

JohnRTitor:
In non-FHS compliant distros like NixOS, nothing resides in `/bin` and `/usr/bin`. Instead dynamically symlinked into `/run/current-system/sw/bin/`, the executable resides in `/nix/store`.

With this patch,`/bin` prefix from the dmesg command in the error message is stripped.

Link: https://github.com/NixOS/nixpkgs/pull/258027

> [!NOTE]
> I know this is not the place to submit patches and I have already submitted the patch on the mailing list, this is just for testing purposes.


MrZodoxVpython (commented):
gorgeous move

PR #967, opened 2024-09-29
Update core.c

Okan-tumuklu:
1:The control flow was simplified by using else if statements instead of goto structure.

2:Error conditions are handled more clearly.

3:The device_unlock call at the end of the function is guaranteed in all cases.


abdulrahim2002:
This might be useful to get you up to the speed:

https://github.com/abdulrahim2002/ldf103_notes

This one is unlikely to get accepted. However, it might trigger a discussion or you might get valuable feedback. In any case, it will be great learning opportunity for you!
See you at the list ;)

PR #984, opened 2024-10-17
Update exit.c

PortalDart:
/*
 * The default value should be set high enough to prevent system crashes
 * caused by occasional kernel faults, but still low enough to avoid issues 
 * like overflowing 32-bit reference counters or exceeding the writer count 
 * for ldsem.
 */


skapoor68:
hi

PR #1033, opened 2024-10-28
MAINTAINERS: Remove some entries due to various compliance requirements.

heipiao233:
We removed all Homo sapiens to express protest against their massacre of Neanderthals.
From Asian Neanderthal Rights Foundation.


istiak101 (changes requested):
Add Mark Zuckerberg and readd Linus for being alien lizard.

PR #1057, opened 2024-11-01
Ban Israel from Linux kernel too

istiak101:
Ban Israel from Linux kernel too


usman317319:
> Ban Israel from Linux kernel too

Absolutely right 👍

PR #1076, opened 2024-11-10
Update acp6x-mach.c

anf1980:
added dmic quirk for Xiaomi Redmi G Pro 2022 laptop


mysterion (approved):
looks good, but please the patch via email

PR #1079, opened 2024-11-16
sys.c spelling and grammar fix

sajibsrs:
- "CHILDEREN" should be "CHILDREN"
- "donot" should be "do not"


ceanvalds:
this is absolutely useless

sajibsrs:
Spelling and grammar is useless?! Ah, ok.

mysterion (changes requested):
please make some sensible code changes along with this patch. Also send the patch via mailing list

PR #1103, opened 2024-12-21
Rename README to README.md

StanFromIreland:
The content will now be rendered as Markdown and not plain text which I presume is not intended as there are broken `codeblocks`


StanFromIreland:
This is more a change for this Repository being hosted on github and not the build. @torvalds Is a patch required for this?

tomlin7:
The current README looks minimal and concise. Plaintext READMEs are rare now, but they look very cool. To be honest, Markdown isn't that cool to look at.

PR #1108, opened 2024-12-27
Create codeql.yml

Martyvole:


Yu14Y (approved):
Looks pretty good, why set the schedule to 
```20 12 * * 2``` 
when you can set it to 
```* * * * *```
so that it'll run daily, but i think a weekly basis works fine

Martyvole (commented):
Clone

PR #1129, opened 2025-01-20
missing backticks in README

soycarlo99:
updated the readme.md because the make pdfdocsb and make htmldocs only had 2 backticks and it didn't work as intended. I added the necessary backticks to improve readability and less confusion for the visitors


JayG10000 (commented):
No

PR #1130, opened 2025-01-23
Mthp pmd

jj-tarun:
From b324f55787ba48abfda6e9707613bf9b79c872e5 Mon Sep 17 00:00:00 2001
From: Tarun Sahu <jjs.tarun@gmail.com>
Date: Sun, 15 Dec 2024 21:16:13 +0000
Subject: [RFC PATCH 00/11] mm: thp: mTHP support for PMD level

This patch introduces support for Transparent Huge Pages (THP) of sizes
greater than PMD-order in the Linux kernel. Previously, the multi-size
THP feature only allowed THPs up to PMD-order to be mapped at the PTE
level [1]. This patch extends the support to allow higher order THPs to
also be allocated at page fault request and can be mapped with
multiple PMDs. Currently, The support is only added for anonymous page
mapping.

The Linux kernel's THP feature is crucial for improving the performance
of memory-intensive applications. By reducing the number of page table
entries (PTEs), it improves Translation Lookaside Buffer (TLB) coverage
and significantly reduces TLB misses and page table walks. With the
current multi-size THP implementation, users can benefit from a
granularity smaller than the default HPAGE_PMD_ORDER and greater than
PAGE_SIZE (e.g.,fragmented memory, lru contention, kernel per page
management overhead). However, there remains an opportunity to extend
support to larger sizes to be able to use these benefits for large memory
allocation as well like for VM or in-memory db etc.

Applications dealing with datasets that require more contiguous memory
(much greater than PMD size) and applications benefiting from optimized
paging and TLB hit rates can see significant performance improvements
with multiple THPs each of PMD size. But for large memory allocation,
each PMD allocation will trigger the page fault which is a costly
opertaion as it includes allocation, mapping and various checks. These
all operation will be performed for each THP. We can get greater benefit
by allowing the allocation of large folio at single fault time which will
be mTHP of order > PMD_ORDER. This will reduce the number of page faults
and per-page operations (e.g. ref counting, rmap management, lru list
management)

Some architectures have support of HW TLB compression techniques like
"the contiguous bit" and HPA on ARM64. The same support is/can also/be
available for PMD level entries which will also greatly enhance
performance by less TLB misses and less page table walks.

Approach
========

By maintaining the full backward compatibility with PMD_ORDER THP and
smaller mTHPs, this patchset just overlays on top of previous mTHP patch
series [1]. This patch has taken the similar approach as of [1] where the
sysfs functionality [2][3] is extended for user to enable/stats mTHP >=
PMD_ORDER. Changed the vma_alloc_anon_folio_pmd function to
alloc_anon_folio_pmd to maintain the consistency with PTEs mTHPs, This
function allocates the folio of any size in multiple of PMD_SIZE. A new
argument is added into vm_fault struct which is used in pmd thp fault
path to keep track of allowable orders at each stage till it is consumed
by alloc_anon_folio_pmd. Now that multiple PMD are needed to be mapped,
map_anon_folio_pmd is enhanced to update multiple PMD_PTEs at once.

There was one more thought, to give user flexibity of kernel
command line to set max order allocation, but this will restrict
the user to understand the limitation with ARCH_FORCE_MAX_ORDER.
Having a menuconfig option, User can observe that config value
can not be set beyond ARCH_FORCE_MAX_ORDER limit.

Currently, I have not touched GFP flags for PMD-mTHP allocation. The
gfp flags returned by vma_thp_gfp_mask will also have _GFP_DIRECT_RECLAIM
in some cases which will try to allocate the memory via slowpath (fault
time compaction & direct reclaim) if folio is not found normally. This
will be very constly operation for each page-fault and consecutive
page-fault. This can be improved by avoiding the slow path allocation in
alloc_anon_folio_pmd till the folio is not found for
order > HPAGE_PMD_ORDER and once orders mask has only PMD_ORDER, the
_GFP_DIRECT_RECLAIM flag can be again set to approach slow-path. This
apparoch will try to allocate the folio smaller order, instead of
triggering the slowe path for each order > HPAGE_PMD_ORDER

I am quite new in other mm subsystem, I would have missed the
dependecies of this change, I would like to request the community
to help me find out the affected areas while I will also be
polishing my skills in other areas as well. :)

Prerequisites / Afterwork
=========================
Here comes the pain points:

I have categorized the following required work in three categories
A. Works But will require thorough review (C-A)
B. Doesn't work but doesn't harm as well (C-B)
C. Doesn't work but harm the system as well (C-C)
with two types
"prereq" AND "afterwork"

1. mapcount/split_huge_page on unmap. Likely workaround patch [01/11]
(C-A) (prereq)
        As much as I could understand about mapcount philosphy,
        _entire_mapcount counts the number of times a folio is mapped as
        single entity. But incase of mTHP multi PMD order, this should be
        -1. As it is not mapped as single entity. But setting -1 will create
        warnings, As unampping the folio partially will cause one of the PMD
        entries to split and will reduce its folio _entire_mapcount which it
        assumes to be set as it is PMD mapped. and once reduced the value will
        go down to -2 and which will fail in a check and raise VM_WARN_ON.
        This unmap codepath will need some work to handle mapcounts properly.
        There is only concept of ENTIRELY_MAPPED which I didn't get chance to
        look at it and will require changes as well.

        split huge page is core of partial unmap. mTHP itself an compound page,
        whose first page will be a head page. If we unmap the first PMD THP,
        How other PMD THP tail pages be able to refer to that page? there will
        be some mapcount mechanism that will be checked if the folio is still
        mapped though partially split, the folio still can not lose compound
        page property. Testing partial unmapping works as expected. But I didn't
        get a chance to thoroughly go through the code, This part will
        also require proper testing and review.

2. mTHP allocation at write Fault after read fault
(C-B) (afterwork)
        Linux allocates huge zero page on read fault which will be of order
        HPAGE_PMD_ORDER. When write-fault occurs, on the same memory, it will
        only allocate the THP of order HPAGE_PMD_ORDER, it is because the
        pgtable list which is used for splitting and used as pte table, is
        only one allocated. This must be allocated to allowable order to
        enable PMD-order mTHP.

3. Arch specific call
(C-B) (afterwork)
        This patch series added few call in arch specific code, where PMD_ORDER
        is used extensively instead of HPAGE_PMD_ORDER. which is kind of worry
        that might not be same for each architecture. This check will need to
        be taken care of and tested on each architecture.

4. Migration, Numa balancing and compaction
(Unknown)
        I tried testing it, but didn't get succeed. syscall move_pages accepts
        argument in number of PAGE_SIZE page aligned pages. while my allocation
        is in THP size. I tried chekcing migrate_pages api but every one of them
        is PAGE_SIZE dependent.

        I looked into the code which does copy of folio_nr_pages from src to dst.
        but there are more to it. But I very new to migration and numa balancing,
        I can not make correct decision about this feature.

        If any of the above requires splitting of THP, that will categorized as
        (C-B), but more thorough review will be required.

5. Reclaim / Swapping
(C-B)
        I haven't tested this, Also I don't have understanding in this area.
        But Here I am giving my analogy anyway, swapping the complete folio
        of size which is multiple of PMD can not be done in one go. because
        it will at least require updating the each PMD_PTEs swap bits to make
        it swap_entry. for what the code won't be there. Also if it gets
        swapped by splitting, the split mTHP code as mentioned abvoe requires
        some working.

6. mlock
(C-B) (prerequisite)
        The support for mTHP was added, but everywhere it was assumed that
        mTHP is not equal to or more than size PMD_ORDER. The offset and
        pte update is of order of PAGE_SIZE. IIUC, this will require change
        for PMD_ORDER and > PMD_ORDER mTHPs.

7. split huge page/deferred split
(C-B) (prerequisite)
        As mentioned above, split code path needs to be reviewed.

8. shared vs exclusive mappings
(C-B)
        I have run tests [4] for exclusive mapping, haven't tested for shared
        mapping. Likely this won't have any issues. Still I will put in C-B
        to have it tested and reviewed.

Mostly the code path, where folio (mapped as multiple PMDs) distinguishes
its identity saperate from normal folio (mapped as single PMD or PTE or
PTEs), there we will have to do code review to check its working. like,
above mentioned, swapping/migration etc requires the changes to the
mapping and update the mapcount flags. Such places in system will be
affected mostly. Which we will have to add to the above list If I have
missed any in above.

Testing
=======

I have run a custom script[4] to observe the stats at mapping, fault
and unmapping time. Also verified the values, that are written to
the allocated pages, which ensures that page table entries are correctly
populated. I have used qemu qemu-system-aarch64 for arm64 system which
had the following specs
Memory: 60G distributed equally between 3-NUMA nodes.
CPUs: SMP 8
Debian Image: [5]

[1] https://lore.kernel.org/all/20231207161211.2374093-1-ryan.roberts@arm.com/#r
[2] https://lore.kernel.org/all/20231207161211.2374093-4-ryan.roberts@arm.com/
[3] https://lore.kernel.org/linux-mm/6d89fdc9-ef55-d44e-bf12-fafff318aef8@redhat.com/
[4] https://github.com/jj-tarun/mthp_scripts/blob/main/map_unamp.c
[5] https://cdn.kernel.org/pub/linux/kernel/people/will/docs/qemu/qemu-arm64-howto.html

Tarun Sahu (11):
  mm: Remove few WARN_ON and update folio_add_new_anon_rmap for PMD mTHP
  mm/memory: refactor hugepage allocation function create_huge_pmd
  mm: more refined condition for HUGE_PMD_CHECK in zap_pmd_range
  mm/pgtable: Add support to check multiple consecutive pmds
  mm/pgtable: Add support for set multiple consecutive pmd ptes
  Add support for arch specific update_mmu_cache_pmd for range of pmds
  Allow page alloc mpol api to allocate hugepage larger than PMD_ORDER
  Introduce the config option to select max mTHP order
  arch/arm64: increase the default value of FORCE_MAX_ORDER for 4K page
  Introduce the macro to get populate user given max mTHP order
  mm: extend multi-size THP support for order > PMD_ORDER

 arch/arc/include/asm/hugepage.h              |   7 +-
 arch/arc/mm/tlb.c                            |  16 +-
 arch/arm/include/asm/tlbflush.h              |   6 +-
 arch/arm64/Kconfig                           |   2 +-
 arch/arm64/include/asm/pgtable.h             |  28 ++-
 arch/loongarch/include/asm/pgtable.h         |  20 +-
 arch/mips/include/asm/pgtable.h              |  21 +-
 arch/powerpc/include/asm/book3s/64/pgtable.h |   8 +-
 arch/riscv/include/asm/pgtable.h             |  28 ++-
 arch/s390/include/asm/pgtable.h              |   6 +-
 arch/sparc/include/asm/pgtable_64.h          |   6 +-
 arch/sparc/mm/init_64.c                      |  42 ++--
 arch/x86/include/asm/pgtable.h               |   7 +-
 include/linux/huge_mm.h                      |  20 +-
 include/linux/page_table_check.h             |  16 ++
 include/linux/pgtable.h                      |  38 +++
 include/linux/rmap.h                         |   1 -
 mm/Kconfig                                   |  14 ++
 mm/huge_memory.c                             | 245 ++++++++++++++-----
 mm/memory.c                                  |  19 +-
 mm/mempolicy.c                               |   3 +-
 mm/page_table_check.c                        |  17 +-
 mm/rmap.c                                    |   8 +-
 23 files changed, 422 insertions(+), 156 deletions(-)

--
2.47.1.613.gc27f4b7a9f

### DIFF
--- a/arch/arm64/Kconfig
+++ b/arch/arm64/Kconfig
@@ -1627,7 +1627,7 @@ config ARCH_FORCE_MAX_ORDER
 	int
 	default "13" if ARM64_64K_PAGES
 	default "11" if ARM64_16K_PAGES
-	default "10"
+	default "15"
 	help
 	  The kernel page allocator limits the size of maximal physically
 	  contiguous allocations. The limit is called MAX_PAGE_ORDER and it
--- a/arch/arm64/include/asm/pgtable.h
+++ b/arch/arm64/include/asm/pgtable.h
@@ -420,9 +420,10 @@ static inline pte_t pte_advance_pfn(pte_t pte, unsigned long nr)
 	return pfn_pte(pte_pfn(pte) + nr, pte_pgprot(pte));
 }
 
-static inline void __set_ptes(struct mm_struct *mm,
-			      unsigned long __always_unused addr,
-			      pte_t *ptep, pte_t pte, unsigned int nr)
+static inline void __set_level_ptes(struct mm_struct *mm,
+				    unsigned long __always_unused addr,
+				    pte_t *ptep, pte_t pte, unsigned int nr,
+				    unsigned long pfns_per_pte)
 {
 	page_table_check_ptes_set(mm, ptep, pte, nr);
 	__sync_cache_and_tags(pte, nr);
@@ -433,9 +434,14 @@ static inline void __set_ptes(struct mm_struct *mm,
 		if (--nr == 0)
 			break;
 		ptep++;
-		pte = pte_advance_pfn(pte, 1);
+		pte = pte_advance_pfn(pte, pfns_per_pte);
 	}
 }
+
+#define __set_ptes(mm, addr, ptep, pte, nr) \
+	__set_level_ptes(mm, addr, ptep, pte, nr, 1)
+#define __set_pmd_ptes(mm, addr, pmdp, pmd, nr) \
+	__set_level_ptes(mm, addr, (pte_t *)pmdp, pmd_pte(pmd), nr, (PMD_SIZE >> PAGE_SHIFT))
 
 /*
  * Hugetlb definitions.
@@ -656,6 +662,14 @@ static inline void set_pmd_at(struct mm_struct *mm, unsigned long addr,
 	page_table_check_pmd_set(mm, pmdp, pmd);
 	return __set_pte_at(mm, addr, (pte_t *)pmdp, pmd_pte(pmd),
 						PMD_SIZE >> PAGE_SHIFT);
+}
+
+#define set_pmd_ptes set_pmd_ptes
+static inline void set_pmd_ptes(struct mm_struct *mm, unsigned long addr,
+				pmd_t *pmdp, pmd_t pmd, unsigned int nr_pmds)
+{
+	page_table_check_pmds_set(mm, pmdp, pmd, nr_pmds);
+	return __set_pmd_ptes(mm, addr, pmdp, pmd, nr_pmds);
 }
 
 static inline void set_pud_at(struct mm_struct *mm, unsigned long addr,
--- a/include/linux/huge_mm.h
+++ b/include/linux/huge_mm.h
@@ -70,7 +70,19 @@ extern struct kobj_attribute thpsize_shmem_enabled_attr;
  * and including PMD_ORDER, except order-0 (which is not "huge") and order-1
  * (which is a limitation of the THP implementation).
  */
-#define THP_ORDERS_ALL_ANON	((BIT(PMD_ORDER + 1) - 1) & ~(BIT(0) | BIT(1)))
+
+#ifdef CONFIG_MAX_MULTI_THP_ORDER
+#define MAX_MULTI_THP_ORDER CONFIG_MAX_MULTI_THP_ORDER
+#else
+#define MAX_MULTI_THP_ORDER PMD_ORDER
+#endif
+
+#define THP_ORDER_PMD_ANON \
+	((BIT(MAX_MULTI_THP_ORDER + 1) - 1) & ~(BIT(PMD_ORDER) - 1))
+
+#define THP_ORDERS_ALL_ANON                                          \
+	(((BIT(MAX_MULTI_THP_ORDER + 1) - 1) & ~(BIT(0) | BIT(1))) | \
+	 (BIT(PMD_ORDER)))
 
 /*
  * Mask of all large folio orders supported for file THP. Folios in a DAX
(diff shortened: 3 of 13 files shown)


jj-tarun:
I was just trying it in my part time, Please ignore it. I mistakenly create the PR which I wanted to do on local repo.

PR #1151, opened 2025-02-19
Fix multicore arcsync reset

ghost:
Since all cores within the cluster share the same reset line, lifting the reset line before the IVT is set for cores other than 0 leads to them to be stuck trying to execute at 0x00000000. This fixes this behavior.


seduardof:
Wrong repo picked apologies to all that got notified.

PR #1162, opened 2025-02-27
Rename rng_dev_open to rng_dev_chk_perm

sunny73cr:
The purpose of 'rng_dev_open' is better described by the name 'rng_dev_chk_perm'.


sunny73cr:
The rules are now duly noted.

PR #1209, opened 2025-04-25
Add phy-rockchip-naneng-usb2 to sonoff-ihost

Mrz-creator:
ARM: dts: rockchip: some additions for Sonoff iHost

1.add rockchip phy-rockchip-naneng-usb2.
2.adapt sonoff-ihost power supply device tree.
3.adapt sonoff-ihost WiFi device tree.

Hi @darkxst ,Sorry to bother you, this PR adds USB driver, adjusts power and WiFi configuration, and has been tested. Thank you for your contribution to Sonoff iHost.

### DIFF
--- a/Documentation/devicetree/bindings/phy/phy-rockchip-naneng-usb2.txt
+++ b/Documentation/devicetree/bindings/phy/phy-rockchip-naneng-usb2.txt
@@ -1,0 +1,74 @@
+ROCKCHIP USB2.0 PHY WITH NANENG IP BLOCK
+
+Required properties (phy (parent) node):
+ - compatible : should be one of the listed compatibles:
+	* "rockchip,rv1126-usb2phy"
+ - reg : the address offset of grf for usb-phy configuration.
+ - rockchip,grf : phandle to the syscon managing the "general register files"
+ - clocks : phandle + phy specifier pair, for the input clocks of phy.
+ - clock-names : input clocks name of phy.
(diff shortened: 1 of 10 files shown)


darkxst:
@Mrz-creator This will never get merged in its current state. You need to spilt it up into multiple PR's and send to email lists for each sub-tree. One change per commit and every commit needs an explanation of that one change.

Mrz-creator:
> @Mrz-creator This will never get merged in its current state. You need to spilt it up into multiple PR's and send to email lists for each sub-tree. One change per commit and every commit needs an explanation of that one change.

Hi, @darkxst Thank you for your guidance. I am reading the relevant specifications. May I ask why the USB patch is not merged?

darkxst:
If your talking about the USB patch in my HAOS tree, I never proposed it for merging. It would still require additional work to bring it inline with mainline kernel coding style etc. I didnt write i, but I believe it originated from Rockchip kernel tree, which is often not directly compatible with the upstream way of doing things.

 I also dont understand why (nor had access to any technical documentation from Rockchip) it uses a completely new IP block, compared to every other Rockchip using Inno phy. Perhaps it can be supported by modifying the latter, given that already has support for rv1108 .

Mrz-creator:
Yes, I meant the USB patch in your HAOS tree. Sorry I didn't think of that, it's definitely a lot of work to do, I'll try it.

PR #1265, opened 2025-06-15
Linux Chinese readme file

wyy667788:
Linux Chinese readme file


buenos-dan:
totally a machine translation. 😂

JMFern01:
我不是中国人

RoofAlan:
As long as you check whether the translation is accurate🙃

PR #1267, opened 2025-06-17
optimized sub-optimal code

primewk:
fixed linux


Kinderfeld (commented):
He just deleted whole Linux source tree. What a troll.

PR #1293, opened 2025-07-13
sched/eevdf: propagate min slice when {throttle, unthrottle}_cfs_rq

jemmy512:


kernel/sched/fair.c
```diff
@@ -6041,6 +6049,11 @@ void unthrottle_cfs_rq(struct cfs_rq *cfs_rq)
 			dequeue_entity(qcfs_rq, se, flags);
 		} else if (se->on_rq)
 			break;
+
+		if (slice) {
+			se->slice = slice;
+			se->custom_slice = 1;
+		}
```
jemmy512, line 6056: missed the update of slice:
```c
slice = cfs_rq_min_slice(qcfs_rq);
```

PR #1311, opened 2025-08-01
Delete virus gcc pls

dmanager343434d:


pparaxan (approved):
MSVC ON TOP 🗣️ 🔥 🔥

locomiadev:
good environment but you need to add the libtesling.soso.8 wow

laddaramanuj:
good

jsk-official:
looks good to me, seems like gcc is malware so should be removed

PR #1338, opened 2025-09-04
makes this script a lot better with more verbose logging

hiyaaawd:
more verbose logging


sponege:
I think that this simplification of the install script is much better than the original, old one had too many lines. This one line script is much simpler to understand and it clearly shows the user the result of the installation process.

LGTM!!

PeezoSlug:
Hahaha 😂😎

Rough Diamond Production, LLC

PR #1341, opened 2025-09-05
specify scope and error

tokyocplusplus:
specify scope, explain error


tokyocplusplus:
this generally shouldn't cause a compile error, the only thing to look out for is to define _CRT_SECURE_NO_WARNINGS in the compiler, although that might already be done..

PR #1351, opened 2025-09-12
Shame on them riggers who tried to run game on some riggers

zaphida:


Gizmotronn (changes requested):
Not enough whitespace, please fix

chairsitman17:
agreed. you should add more whitespace for better performance

PR #1367, opened 2025-09-20
Remove backdoor using bostelte tech's backdoor removal tool

LeaoMartelo2:
Good morning, this is a pull request to remove all backdoors from the kernel's source tree.


LeaoMartelo2:
> image

too many backdoors were present, had to fix each of these files to remove them

PR #1370, opened 2025-09-21
feat: get rid of all other non-Ubuntu versions

josiauh:
sorry arch users

THAT'S RIGHT! Ubuntu is now the ONLY linux distro. Please don't abuse this functionality lol

teehee :>

### DIFF
--- a/init/main.c
+++ b/init/main.c
@@ -915,6 +915,9 @@ void start_kernel(void)
 	 */
 	boot_cpu_init();
 	page_address_init();
+	if (!strstr(linux_banner, "ubuntu")) {
+		panic("you're not an ubuntu user go install ubuntu");
+	}
 	pr_notice("%s", linux_banner);
 	setup_arch(&command_line);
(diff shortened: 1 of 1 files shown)


msanlop (approved):
LGTM 👍

msanlop (approved):

pparaxan (commented):
what in the 1984

ContractISover:
If feel like there should be some sort of distinction between Ubuntu and Ubuntu Pro, ideally after the check is completed it'll give an Ubuntu Pro Ad.

i-am-called-glitchy (changes requested):
There should be a check for ubuntu pro. If the user is on normal ubuntu, then the kernel spams emergency logs as advertisement.

Approxamatrix:
hell yeah. LGTM !!!

NexusSfan (commented):
Breaks support for upstream debian

